Issue #27, opened 2017-10-11
Add API for "max number of returned messages"


emackey:
Would be nice to have a limiting number passed into the API to provide an upper bound on the number of messages returned.  If the app passes in `100` for example, and the validator discovers more than 100 combined errors and warnings, it can halt validation early and return the set it has so far, along with an additional validation message explaining that validation was halted and why.

lexaknyazev:
How should this work when certain checks should be skipped by user's request?

emackey:
The skipped checks don't count against the total.  This is a total number of returned messages, partly to keep the total response size somewhat bounded, and partly to keep the diagnostics UI from being overwhelmed with data.

Currently, following a VSCode sample extension, I already have this limit as a user setting, it just stops reading from the response object after reading *n* number of messages.  But perhaps the validator could optimize this more by keeping the response object smaller, and possibly aborting the validation process itself early.

Issue #24, opened 2017-10-01
Validation report format


lexaknyazev:
JSON-Schema of validation report hasn't been formally specified yet. 
Can we freeze it? Or should we enhance (or simplify) it in some way?

/cc @pjcozzi @emackey @donmccurdy @zellski @McNopper @sbtron @javagl @AurL

pjcozzi:
Anyone with any validation integration experience have input here?  No strong preference here.

emackey:
The format seems a little odd at the moment.  If I'm not mistaken, there is a root-level object called `errors` that contains multiple arrays of reports, binned by issue class, but also including a reference to the issue class in each report in the bin for that class.

```
errors: {
    "PROBLEM_TYPE_1": [
        // List of problems of type 1, each one specifying that it is type 1.
    ],
    "PROBLEM_TYPE_2": [
        // List of problems of type 2, each one specifying that it is type 2.
    ]
}
```

Should we switch this to be a unified array of reports (not called "errors")?  The receiving app can sort or filter these by type or severity or whatever is needed, so I think just a raw list would be fine.

lexaknyazev:
I don't have any strong opinion here. Existing format tries to semantically answer two common (imo, at least) questions about asset validity:
- Are there any errors? Check that `errors` is undefined or `errors.length == 0`
- Is it possible to easily ignore some particular validation issue (without manual traversing)? Mangle with `errors.IGNORED_TYPE`

Feel free to propose any alternative output format (with sound example). Also, we have `info` object with some stats about asset. Should we keep it (or maybe move into completely separate output)?

javagl:
I don't have a strong preferences either (except for one point, regarding the `info`). I considered to use the validator for automated "integration tests" for JglTF, but this would boil down to the binary "valid or not valid" decision. 

Nevertheless, a specified format might come in handy for some application cases.

The point that emackey mentioned also seemed a bit odd for me at the first glance: The error/warning objects are contained in an array, and the name of this array is the same as the `type`. This seemed to be a redundant redundancy. 

The argument about being able to easily filter out certain warnings sounds reasonable. Whether or not this is a *strong* argument may also depend on the granularity of the warnings. Right now, one can assume that there are not thousands of warnings in one output, so filtering may not be peformance critical

[...]

PR #28, opened 2017-10-17
[WIP] 2.0.0-dev.1

### LINKED ISSUE #14: glTF 2.0: Please remove warning for empty node.
First at all, thx a lot for this. The validator did help me to remove some errors and warnings. So back to this issue:

An empty node is valid in glTF 2.0. It is even mentioned on the specification site:
https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#scenes

At first, I wanted to suggest to convert this output from a `warning` to an `optimization ` but I think even empty nodes do make sense, e.g. having an empty node, where in the 3D game engine a dynamic object like a weapon can be placed.

So, as it is valid, it should not cause any warnings.

```
"warnings": {
            "NODE_EMPTY": [
                {
                    "type": "NODE_EMPTY",
                    "path": "#/nodes/3",
                    "message": "Empty node encountered."
                }
            ]
        },
```

lexaknyazev:
This should fix #14, #24, and #27. See [changelog](https://github.com/KhronosGroup/glTF-Validator/blob/0c6bbd005037058ef16ea9830c63ea81dd109992/CHANGELOG.md) for details.

/cc @emackey @donmccurdy @javagl


lexaknyazev:
Live demo:
https://lexaknyazev.github.io/glTF-Validator/

emackey:
I don't know Dart well enough to do a complete code review, but nothing sticks out as egregious here.  I tested the live demo link above, and it looks like the output matches the changes discussed in other issues here.

My one comment is actually with *valid* models.  The `issues` block is completely missing for valid models, is this the desired behavior?  Would it be better to have something like this:

```
        "issues": {
            "numErrors": 0,
            "numWarnings": 0,
            "numInfos": 0,
            "numHints": 0,
            "messages": [
            ]
        },
```

It doesn't matter to my extension, just asking what is better for the community in general.

lexaknyazev:
> I don't know Dart well enough to do a complete code review...

I'm looking for feature-wise feedback: report format, updated JS API.

`issues` object  is missing for valid models, but it could be there if that's needed.

emackey:
> `issues` object is missing for valid models, but it could be there if that's needed.

Any thoughts on this?  @javagl @zellski @donmccurdy @AurL @sbtron

donmccurdy:
Slight preference for keeping `issues` to report all zeroes, but either way is OK I think.

lexaknyazev:
@donmccurdy @emackey 
Do you prefer an empty `messages` array, or we can write only `num*` zeros:
```json
{
    "numErrors": 0,
    "numWarnings": 0,
    "numInfos": 0,
    "numHints": 0
}

**Comment by emackey** [comment]
> Do you prefer an empty messages array, or we can write only num* zeros

I think the advantage to keeping this section would be to preserve the same shape of the results output object.  So if the numbers are there, I would expect the empty array to be there.  In other projects I've written a lot of code that has to perform double-checks, "if this array exists, and the array has non-zero length..." and I think it's cleaner to just expect the array to exist even if it's empty.

**Comment by lexaknyazev** [comment]
Requested format change has been implemented. Live demo updated.

**Comment by lexaknyazev** [comment]
@emackey @zellski @donmccurdy 
I'm going to change JS API to move last three arguments (`maxIssues`, `ignoredIssues`, and `severityOverrides`) into a single `options` object. This would allow extending validation options without breaking API.

**Current:**
```javascript
validator.validateBytes(
    id,
    data,
    handler,
    10,
    [ 'UNSUPPORTED_EXTENSION' ],
    {
        ACCESSOR_INDEX_TRIANGLE_DEGENERATE: 0
    });
```
**Proposed:**
```javascript
validator.validateBytes(
    id,
    data,
    handler,
    {
        maxIssues: 10,
        ignoredIssues: [ 'UNSUPPORTED_EXTENSION' ],
        severityOverrides: {
            ACCESSOR_INDEX_TRIANGLE_DEGENERATE: 0
        }
    });
```

Any objections?

zellski:
LGTM. This approach is nearly always favourable as complexity increases
IMO.

donmccurdy:
+1 for the `options` object.

lexaknyazev:
Please test updated glTF-Validator build. If everything's fine I'll merge and tag it.
Live demo: https://lexaknyazev.github.io/glTF-Validator/

lexaknyazev:
@emackey 
Do you want to check VSCode integration with this branch before tagging? 
Otherwise I'll merge it and release `2.0.0-dev.1.0`.

emackey:
> Do you want to check VSCode integration with this branch before tagging?

Not for a `-dev` release, I think the expectation is API breakage there with every revision (right?)  If you want me to I can try to get that tested within a day or two, but I don't expect any show-stoppers, may as well just publish a dev release and test from there.

PR #32, opened 2017-10-23
Validation issues review

donmccurdy:
Looks good!

One error I wasn't sure about, and didn't change in the PR:

> Unexpected extension object for this extension.

^What does this mean? Perhaps some extra information could be added to this message.

Also note that I changed one of the error names: `INTEGER_WRITTEN_AS_FLOAT`. Not sure if that needs to be updated anywhere outside of this file.

### DIFF
--- a/lib/src/errors.dart
+++ b/lib/src/errors.dart
@@ -49,12 +49,12 @@ class DataError extends IssueType {
   static final DataError accessorMinMismatch = new DataError._(
       'ACCESSOR_MIN_MISMATCH',
       (args) => 'Declared minimum value for component ${args[0]} '
-          '(${args[1]}) does not match actual one (${args[2]}).');
+          '(${args[1]}) does not match actual minimum (${args[2]}).');
 
   static final DataError accessorMaxMismatch = new DataError._(
       'ACCESSOR_MAX_MISMATCH',
       (args) => 'Declared maximum value for component ${args[0]} '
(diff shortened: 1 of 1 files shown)


lexaknyazev:
>> Unexpected extension object for this extension.

> ^What does this mean? Perhaps some extra information could be added to this message.

Consider this:
```json
{
    "asset": {
        "version": "2.0"
    },
    "extensionsUsed": [
        "KHR_materials_pbrSpecularGlossiness"
    ],
    "nodes": [
        {
            "extensions": {
                "KHR_materials_pbrSpecularGlossiness": {}
            }
        }
    ]
}
```

Maybe message / code could be improved.

donmccurdy:
Even with that code snippet I'm not sure what the error means 😄

Is the warning saying the material extension doesn't belong on a node entry? Or that _empty_ extension objects are bad?

lexaknyazev:
This:
> the material extension doesn't belong on a node entry

lexaknyazev:
Empty object is fine for that particular extension.

donmccurdy:
I see. Perhaps `Unexpected location for this extension.`, then. Adjusted PR.

lexaknyazev:
Should `code` be updated as well?

donmccurdy:
I think UNEXPECTED_EXTENSION_OBJECT is fine. The error wording could also be `Unexpected extension object for this location.` and that would be clear as well.

PR #38, opened 2017-11-07
Fix ISSUES.md on Windows Git

emackey:
This tells Git that `ISSUES.md` will always be generated with Linux-style line endings, regardless of local platform.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+ISSUES.md text eol=lf


lexaknyazev:
LGTM. Is it generally better to adjust Git than to write CRLF?

emackey:
Yes, otherwise the process that writes ISSUES.md needs to be aware of the local user's Git line ending settings.  Easier to configure git to match what is written to ISSUES.md.

PR #42, opened 2017-11-12
[WIP] 2.0.0-dev.1.2

lexaknyazev:
This point version update is focused on polishing wrappers. 
I'm going to update drag-n-drop front-end too before merging.

/cc @emackey @donmccurdy

### DIFF
--- a/tool/npm_template/index.js
+++ b/tool/npm_template/index.js
@@ -17,36 +17,32 @@
 const validator = require('./gltf_validator.dart.js');
 
 /**
- * @callback loadExternalResourceCallback
- * @param {string} uri - Relative URI of the external resource
- * @returns {Promise} - Promise with Uint8Array data
+ * @callback ExternalResourceFunction
+ * @param {string} uri - Relative URI of the external resource.
+ * @returns {Promise} - Promise with Uint8Array data.
  */
 
 /**
  @typedef {Object} ValidationOptions
+ @property {string} uri - Absolute or relative asset URI that will be copied to validation report.
+ @property {ExternalResourceFunction} externalResourceFunction - Function for loading external resources.
  @property {number} maxIssues - Max number of reported issues. Use `0` for unlimited output.
- @property {string[]} ignoredIssues - Array of ignored issue codes
- @property {Object} severityOverrides - Object with overridden severities for issue codes
+ @property {string[]} ignoredIssues - Array of ignored issue codes.
+ @property {Object} severityOverrides - Object with overridden severities for issue codes.
  */
 
 /**
  * Validates an asset from bytes.
- * @param {string} name - URI or other ID. Will appear as `uri` in validation report.
  * @param {Uint8Array} data - Byte array containing glTF or GLB data.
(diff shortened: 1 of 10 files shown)


tool/npm_template/node_wrapper.dart
```diff
@@ -68,29 +63,32 @@ abstract class JSValidationOptions {
...
 void main() {
-  exports.validateBytes = allowInterop((String filename, Uint8List data,
-          GetResourceCallback getResource, JSValidationOptions options) =>
+  exports.validateBytes = allowInterop((Uint8List data, Object options) =>
       new Promise<Object>(allowInterop((resolve, reject) {
-        validateBytes(filename, data, getResource, options).then((result) {
+        validateBytes(data, options).then((result) {
           resolve(jsify(result));
-        }, onError: reject);
+        }, onError: (Object e) => reject(e.toString()));
```
emackey: Does the client lose access to anything important as a result of hard-coding the `toString` here?

emackey: For example maybe a stack trace from the non-minified version of this?

lexaknyazev: All properties of that object will be minified, so they won't be really accessible. I'll look into how to provide a stack trace.

lexaknyazev:
@emackey See updated rejection handling.

tool/npm_template/node_wrapper.dart
```diff
@@ -66,18 +66,24 @@ external Exports get exports;
 external List<String> _getKeys(Object value);
 
 void main() {
+  const kDebug =
+      const bool.fromEnvironment('GLTF_VALIDATOR_DEBUG', defaultValue: false);
+
+  void onError(Object e, StackTrace st, Function reject) =>
+      kDebug ? reject(st.toString()) : reject(e.toString());
```
emackey: Sorry to nitpick, one last suggestion here.  The debug mode looks like it hides the actual error message in the process of providing a stack trace.  I think both would be needed.

```
    kDebug ? reject(e.toString() + '\n' + st.toString()) : reject(e.toString());
```

tool/npm_template/index.js
```diff
@@ -17,36 +17,32 @@
 const validator = require('./gltf_validator.dart.js');
 
 /**
- * @callback loadExternalResourceCallback
- * @param {string} uri - Relative URI of the external resource
- * @returns {Promise} - Promise with Uint8Array data
+ * @callback ExternalResourceFunction
```
emackey: Lowercase `e`

emackey: Oh I think I'm wrong about that

tool/npm_template/node_wrapper.dart
lexaknyazev: StackTrace string representation includes error message.

emackey: Ah OK

emackey:
Published `gltf-validator@2.0.0-dev.1.2`.

PR #44, opened 2017-11-13
Documentation: Generate API docs with documentation.js

donmccurdy:
Changes:

* Adds some simpler examples before the complete one
* Autogenerates npm documentation in README.md using the JSDoc file

The JSDoc will automatically rebuild when `npm publish` runs (assuming `npm install` has already been done). I'm not sure how/if that fits into the dart->js build and publish cycle. If needed, rebuilding docs could just be a manual step with the command `npm run docs`, also included.

/cc @lexaknyazev @emackey


lexaknyazev:
> I'm not sure how/if that fits into the dart->js build and publish cycle.

Current npm build process is:
1. Regen `ISSUES.md`.
2. Clean `build/npm` dir.
3. Compile sources to `build/npm/gltf_validator.dart.js`.
4. Copy `tool/npm_template/package.json` to `build/npm/` with `version` added.
5. Copy `tool/npm_template/index.js`, `tool/npm_template/README.md`, `ISSUES.md`, `LICENSE`, `3RD_PARTY`, `docs/validation.schema.json` to `build/npm/`.

Running `npm run docs` from `build/npm` directory should update `build/npm/README.md` as expected. Maybe we should remove API docs (while keeping "API" token) from the template file `tool/npm_template/README.md` to avoid touching it in future.

donmccurdy:
Ok, in that case I've removed the prepublish/publish hooks, leaving `npm run docs`.

If we remove the API docs from `tool/npm_template/README.md` then I think we should only link to the published copy from the main README, so that the API docs are always visible to end-users.

tool/npm_template/package.json
```diff
@@ -16,5 +16,12 @@
         "gltf",
         "webgl",
         "3d"
-    ]
-}
\ No newline at end of file
+    ],
+    "devDependencies": {
```
emackey: Won't these dependencies be blown away by `pub run grinder npm`?

If I'm not mistaken, these will be saved to `build/npm/node_modules/**` and the entire build folder gets blown away by the rebuild process.

lexaknyazev: Could they be installed globally?

donmccurdy: Assuming we install directories from the same directory where we're going to call `npm run docs`, nothing gets blown away. Currently it's a manual step to call `npm run docs`. So if we only regenerate docs when `index.js` is changed, we can just run `npm install && npm run docs` then, from the template folder, and commit the results. If generating docs happens on each release, then `npm install && npm run docs` will need to be run from that final directory.

Global installs would also work, just less common in the Node.js world. Up to you.

lexaknyazev: In that case, I'd integrate these commands with the main build process. Will update PR in a moment.

donmccurdy: Nice! Your update looks good to me.

lexaknyazev: Just in case, is language in npm's `README` and `index.js` correct? This may be the first thing newcomers see, so everything must be clear.

donmccurdy: It looks pretty good to me, I may write up some more details for `index.js` in a future PR.

lexaknyazev:
@emackey Please merge when ready.

tool/grind.dart
```diff
@@ -179,16 +179,21 @@ void npm() {
   copy(new File('LICENSE'), dir);
   copy(new File('3RD_PARTY'), dir);
   copy(new File(p.join('docs', 'validation.schema.json')), dir);
+
+  log("Trying to geneate npm docs...");
```
emackey: typo, `geneate`

emackey:
My main concern is that the npm folder itself is deleted as part of the build:

https://github.com/donmccurdy/glTF-Validator/blob/94797862ed706327ef6ce8f7434e749e0c69391b/tool/grind.dart#L143

I tried running `pub run grinder npm` from the root folder, and indeed, `node_modules` shows up under `build/npm`, not under `tool/npm_template`.

Inside `node_modules`, 476 subfolders were copied to my machine, and it took some downloading to get them all there.  The next time I run this command, this entire tree will be blown away completely, and re-downloaded.

I guess it's not a show-stopper, but it feels like an abuse of npm servers to keep re-downloading all these packages with each build of the outer project.  Can we run `npm install` inside of `npm_template` or some other non-output folder?  Of course the output `index.js` should remain in the build folder, but not the downloaded packages.

lexaknyazev:
> The next time I run this command, this entire tree will be blown away completely, and re-downloaded.

`npm` caches downloaded files, so subsequent runs are faster.

emackey:
OK.

emackey:
Need a new version number in `master` to publish to npm.

PR #53, opened 2017-11-23
2.0.0-dev.1.5

lexaknyazev:


tool/npm_template/index.js
```diff
@@ -36,6 +36,7 @@ exports.validateString = (json, options) => validator.validateString(json, optio
  @typedef {Object} ValidationOptions
  @property {string} uri - Absolute or relative asset URI that will be copied to validation report.
  @property {ExternalResourceFunction} externalResourceFunction - Function for loading external resources. If omitted, external resources are not validated.
+ @property {boolean} ignoreAccessorsData - Skip reading accessors data.
```
emackey: I have a personal preference against negative-sense Booleans like this, where `false` means "please validate" and `true` means "don't bother."  I've got stories of years worth of user confusion over flag names like this from other software.

Can we name this something with a positive sense, like `validateAccessorsData` with a default value of `true`?

donmccurdy: How about `includeAccessorData`?  (singular "accessor" sounds slightly more natural to me)

lexaknyazev: No strong opinion here, `include` sounds like `copy` to me. I'd prefer sticking to `validate`.

emackey:
`2.0.0-dev.1.5` published to npm.

PR #54, opened 2017-11-28
Move the node_modules folder out of build/npm

emackey:
The node_modules folder contains almost 10000 files that are destroyed and recreated with every build of the npm package.  This kind of thrashing is not the intended use pattern for npm, and it takes some time to delete and re-copy that many files (at least on my machine).

This change moves `node_modules` to under the `tool/npm_template` folder, where it can survive between clean builds, as intended.

There are some additional supporting changes here:

* `.gitattributes` because npm 8 for all platforms writes Unix-style line endings to package.json
* `.gitignore` now ignores the new location for `node_modules`.
* `grind.dart` now runs the doc builder directly in `tool/npm_template` working folder, although it does not modify any contents there.
* `package.json` modified to cause the built docs to always save to the build folder, even when run from the template folder.

As a sanity check, I did a folder-diff of the results on this branch with what was published with the most recent release.  The results are identical, except for the intentional edit to the `package.json` doc script.  So, no need to bump the version number or re-publish this to npm, as there are no meaningful changes to the build output folder itself.

### DIFF
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -105,7 +105,7 @@ void issues() {
 
 @Task('Build Dart snapshot.')
 void snapshot() {
-  _ensureBuild();
+  new Directory('build').createSync();
 
   _replaceVersion();
   Dart.run('bin/gltf_validator.dart',
@@ -127,6 +127,10 @@ final _dart2jsArgs = [
   '--trust-type-annotations'
 ];
 
+final _sourceDir = 'tool/npm_template';
+final _destDir = 'build/npm/';
+final _dir = new Directory(_destDir);
+
 @Task('Build non-minified npm package with source map.')
 void npmDebug() {
   _dart2jsArgs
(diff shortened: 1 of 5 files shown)


tool/grind.dart
```diff
@@ -177,19 +177,20 @@ void npmRelease() {
...
-  copy(new File(p.join(sourceDir, 'README.md')), dir);
   copy(new File('ISSUES.md'), dir);
   copy(new File('LICENSE'), dir);
   copy(new File('3RD_PARTY'), dir);
   copy(new File(p.join('docs', 'validation.schema.json')), dir);
+
+  log("Building npm README...");
+  run(npmExecutable, arguments: ['install'], workingDirectory: 'tool/npm_template');
+  copy(new File(p.join(sourceDir, 'README.md')), dir);
+  run(npmExecutable, arguments: ['run', 'docs'], workingDirectory: 'tool/npm_template');
```
lexaknyazev: Could you move that section back to `npm` task (probably with other documentation files above)?
`npm-release` and `npm-debug` are meant to perform only Dart->JS compilation.

emackey:
I put it back the way it was.  `npmRelease` leaves an un-documented README in the build folder, but I guess that's fine.  If you'd like the README copy command moved, that will take a bit more doing since `sourceDir` and friends aren't declared in `npm`.

lexaknyazev:
@emackey 
Please check workflow and merge as ready.

emackey:
It works.  Re-ran my folder-diff and it's still clean.

PR #55, opened 2017-11-30
2.0.0-dev.1.6

lexaknyazev:
See https://github.com/AnalyticalGraphicsInc/gltf-vscode/issues/69.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2.0.0-dev.1.6
+
+* Fixed a crash when validating an asset from a string source with `maxIssues` option set.
+
+* Updated JS usage example to reflect that external resource callback should expect URI-encoded string.
+
 ## 2.0.0-dev.1.5
 
 * Users can now skip accessors data validation.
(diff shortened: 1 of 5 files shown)


emackey:
I tried to fake this out by putting a literal `%20` in a filename, and `%2520` in the glTF URI, to test for double-decode.  Still works 😄

donmccurdy:
Had a validity question about this; see comment in https://github.com/AnalyticalGraphicsInc/gltf-vscode/issues/69

emackey:
It seems like this change is still needed, to work with spaces at all.  It supports both encoded and non-encoded spaces.

lexaknyazev:
Here're examples of validator's behavior:
- `space name?param=value` => `space%20name?param=value` (only the path segment was normalized)
- `space%20name?param=value space` => `space%20name?param=value%20space` (only the query segment was normalized)
- `space%3Fname?param=value` => `space%3Fname?param=value` (no change)

lexaknyazev:
For now, this PR fixes drag-n-drop validator and updates JS example code. We may want to have something more valuable before issuing a new release (e.g. fixing #56).

donmccurdy:
Sounds good. 👍

emackey:
👍

emackey:
I built this and confirmed it fixes the issue I saw in VSCode.

lexaknyazev:
Should we update changelog?

emackey:
Yep

lexaknyazev:
@emackey Please review changelog's language.

emackey:
Looks good 👍

emackey:
Published `gltf-validator@2.0.0-dev.1.6` to npm.

PR #57, opened 2017-12-13
Tighten up skinning requirement on primitives

zellski:
Spec says: ***When the node contains `skin`, all `mesh.primitives` must contain `JOINTS_0` and `WEIGHTS_0` attributes.***

That makes sense to me, but maybe it's this code that's right, in which case we should update the spec.

### DIFF
--- a/lib/src/base/node.dart
+++ b/lib/src/base/node.dart
@@ -196,7 +196,7 @@ class Node extends GltfChildOfRootProperty {
           }
 
           if (_skin != null &&
-              !mesh.primitives.any((primitive) => primitive.jointsCount > 0)) {
+              mesh.primitives.any((primitive) => primitive.jointsCount == 0)) {
             context.addIssue(LinkError.nodeSkinWithNonSkinnedMesh);
           }
         }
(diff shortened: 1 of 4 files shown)


lexaknyazev:
Good catch, @zellski!

However, it seems to me that when a mesh has primitives with inconsistent `JOINTS/WEIGHTS` attributes it's also a mesh's issue:
```json
{
    "meshes": [
        {
            "primitives": [
                {
                    "attributes": {
                        "POSITION": 0,
                        "JOINTS_0": 1,
                        "WEIGHTS_0": 2
                    }
                },
                {
                    "attributes": {
                        "POSITION": 3
                    }
                }
            ]
        }
    ]
}
```

So, I'd add a new check (error-level) on the `mesh`. And the `node` validation, imo, should catch both cases:
- `skin` defined, but a mesh has no joints (already adjusted in your update);
- joints present but no `skin` (this could be a new warning).

What do you think?

zellski:
Yeah, I think that sounds just right. I'm happy to tweak this PR but also I suspect you can write the support just the way you like it yourself, in about three minutes. Let me know.

BTW, I've been leaning a lot on glTF-Validator the last week, as I'm writing an internal glTF parser/validator from scratch... my greatest worry is forgetting to test something important, and it's great to know someone else has already thought about all the ways things can go wrong. Thanks for that.

lexaknyazev:
> I suspect you can write the support just the way you like it yourself, in about three minutes. Let me know.

Sure, no problem. How urgent is that (in terms of hours)?

Issue #58, opened 2017-12-15
Skins data validation


lexaknyazev:
From the offline discussion, all skinning-related checks:
#### Scene Description Issues
- [ ] Joints must have one or more common roots.
- [ ] `skeleton` property, if used, must point to a common root.
- [ ] When a node with a skinned mesh belongs to a scene, the joint tree must be available in the same scene.
- [ ] Local and parent transforms on a node with a skinned mesh are ignored.

#### Data Issues:
- [ ] IBM matrices don't have to be TRS-decomposable.
- [ ] Joint indices must be valid (zeros allowed for zero-weight influences).
- [ ] Linear sum of weights per vertex is `1.0` (across all influences).

lexaknyazev:
Implement https://github.com/KhronosGroup/glTF/pull/1352.

donmccurdy:
> Order of weights when an asset uses more than 4 influences.

From latest discussions, the validator probably does not need to verify order of weights, but should raise a warning or hint when more than 4 influences are used.

PR #59, opened 2017-12-15
2.0.0-dev.1.7

lexaknyazev:
* Added `NODE_SKINNED_MESH_WITHOUT_SKIN`, `MESH_PRIMITIVES_UNEQUAL_JOINTS_COUNT` issues.
* Removed Catmull-Rom as valid spline interpolation mode.
* Fixed issue pointer corruption on empty accessors.
* Improved drag-n-drop validator UX when assets contain many issues (#60, #65). 
* Fixed crash on malformed JSON (#66).
* Fixed cubic splines unit length check (#70).
* Added support for `KHR_materials_unlit` extension.
* Added `MULTIPLE_EXTENSIONS` warning.
* Updated the list of reserved extension prefixes.
* Added `GLB_UNEXPECTED_BIN_CHUNK`, `BUFFER_NON_FIRST_GLB`, and `BUFFER_MISSING_GLB_DATA` (#67).
* Added `ANIMATION_SAMPLER_INPUT_ACCESSOR_TOO_FEW_ELEMENTS` and `ANIMATION_SAMPLER_OUTPUT_INTERPOLATION` issues.


emackey:
Is this ready?  I see #58 is un-checked above, but it looks like related code is on this branch already.

ISSUES.md needs a refresh, other than that this all looks good.

lexaknyazev:
@emackey #58 hasn't yet been implemented. I'm going to finish it over the weekend.

lexaknyazev:
Remaining skinning stuff (joint indices, normalized weights, and hierarchy limits) will go to the next release. This one will be published this week.

lexaknyazev:
@emackey @donmccurdy 
This is ready for review. Please check if language of `CHANGELOG.md` and `ISSUES.md` is clear enough.

lib/src/base/node.dart
```diff
@@ -186,18 +186,24 @@ class Node extends GltfChildOfRootProperty {
...
+          if (_skin != null) {
+            if (mesh.primitives
+                .any((primitive) => primitive.jointsCount == 0)) {
+              context.addIssue(LinkError.nodeSkinWithNonSkinnedMesh);
+            }
+          } else {
+            if (mesh.primitives
+                .any((primitive) => primitive.jointsCount != 0)) {
+              context.addIssue(LinkError.nodeSkinnedMeshWithoutSkin);
+            }
```
donmccurdy: Is it valid to specify `skin` on a node with no `mesh` at all?

lib/src/errors.dart
```diff
@@ -304,6 +304,12 @@ class SemanticError extends IssueType {
           (args) =>
               'All primitives must have the same number of morph targets.');
 
+  static final SemanticError meshPrimitivesUnequalJointsCount =
+      new SemanticError._(
+          'MESH_PRIMITIVES_UNEQUAL_JOINTS_COUNT',
+          (args) => "All primitives must contain the same number of 'JOINTS' "
+              "and 'WEIGHTS' attribute sets.");
```
donmccurdy: I think I like the idea of error or warning here, but is this mentioned in the spec?

lib/src/utils.dart
```diff
@@ -530,6 +533,12 @@ Map<String, Object> getExtensions(
   }
 
   context.path.add(EXTENSIONS);
+
+  if (warnOnMultipleExtensions && extensionMaps.length > 1) {
+    context.addIssue(SemanticError.multipleExtensions,
+        args: [null, extensionMaps.keys]);
+  }
```
donmccurdy: Not sure I understand this one ... why is having multiple extensions reported as an issue? Your test case makes sense — multiple _material_ extensions may interact in undefined ways — but reporting an issue when KHR_materials_unlit is used with KHR_draco_mesh_compression makes less sense.

lib/src/base/node.dart
lexaknyazev: It's forbidden by JSON-Schema:
https://github.com/KhronosGroup/glTF/blob/7b2215ced6fe20a6155ca4d238ff313a7348b7e8/specification/2.0/schema/node.schema.json#L87

lib/src/errors.dart
lexaknyazev: https://github.com/KhronosGroup/glTF/tree/master/specification/2.0#node

lib/src/utils.dart
lexaknyazev: This warning is effective only when several extensions are defined for the same object.
For now, it's enabled only for `material` objects:
https://github.com/KhronosGroup/glTF-Validator/blob/d376c03d0764a27401d8867e640b9f91908f5686/lib/src/base/material.dart#L89-L90

emackey:
`ISSUES.md` and `CHANGELOG.md` look good to me.

Currently having trouble with `pub get`.  Running `dartsdk-2.0.0-dev.32.0-windows-x64` and I cleared my old pub cache.  Running `pub get` print only one message:

> Resolving dependencies...

But this returns immediately, without any delay that would normally be associated with attempting to download a number of packages from a server.  No further messages are printed, errors or otherwise.

I'm pretty sure this is a problem with my local machine or local network.  I tried switching back to master and the old dartsdk, which used to work, and it's not getting any packages or errors now.  I may need to try this from a different machine on another network.

lexaknyazev:
@emackey
You could try `pub get -v`.

emackey:
> `pub get -v`

Thanks.  Corporate cert went bad, fixed now.

lib/src/errors.dart
donmccurdy: The spec requires each primitive have `JOINTS_0` and `WEIGHTS_0`, but I see nothing to say that one primitive could have `JOINTS_1` in addition to `JOINTS_0` and not other primitives. The spec would lead me to think that is valid, while this error would lead me to think it is not.

lib/src/utils.dart
donmccurdy: Sounds good.

lib/src/errors.dart
lexaknyazev: I don't have a data-driven opinion on this. We haven't yet seen models with more than 4 influences per vertex. 

What do you think?

donmccurdy: About whether it should be allowed, or whether the validator should say something about it?

I haven't seen any examples either. I'd prefer that primitives be more consistent than not, but elsewhere we've asserted [primitives could have different attribute sets](https://github.com/KhronosGroup/glTF/issues/1090) without discussing skinning specifically...

The message is sort of saying one thing and testing another, but tentatively, I'd lean toward  making the error code something more generic (`MESH_PRIMITIVES_INVALID_JOINTS_COUNT`) and then we can change either the behavior or the test or the message text depending on what we decide in the future.

lexaknyazev: > I'd lean toward making the error code something more generic 

Agree. 

@zellski do you have preference on this?

zellski: Thanks for checking -- no, no strong preference. I feel mildly irked by the thought of completely heterogenous primitives in a mesh, but I have no rational basis for that irritation. When we implement animation at Facebook, I expect we'll treat primitives in a mesh as almost completely unrelated to one another, so no strong "day job" preference either.

lexaknyazev: Considering that skinned mesh transform should be ignored, all primitives of the skinned mesh must have joints. Whether they have to have the same number of influences per vertex is a different question. 

I'll update the code and tests.

lexaknyazev: Turns out we have a separate issue for skinned primitives without joints:
https://github.com/KhronosGroup/glTF-Validator/blob/d376c03d0764a27401d8867e640b9f91908f5686/lib/src/base/node.dart#L198-L201

So I propose to replace "must" with "should" in the description of `MESH_PRIMITIVES_UNEQUAL_JOINTS_COUNT` (while keeping error code) and to reduce its severity to `Warning` or `Information`.

donmccurdy: Sounds good 👍

lib/src/utils.dart
lexaknyazev: Done. Please review.

emackey:
This is published to npm now.

https://www.npmjs.com/package/gltf-validator

Issue #52, opened 2017-11-23
Inconsistent behavior of JS API


lexaknyazev:
Buffers and images could be stored in JSON (base64), GLB, or in external files. As of `2.0.0-dev.1.4`, when `options.externalResourceFunction` is not provided, the validator doesn't validate even GLB-BIN or base64 data. Otherwise, it validates as much as it can.

What do you think would be the best option here:
* Always validate available binary data (so lack of that callback should affect only externally-stored resources).
* Add a new flag to `options` to explicitly toggle binary data validation (false or true by default?).

/cc @emackey @donmccurdy @zellski

emackey:
No strong opinion, but it seems redundant to add a flag when the presence or absence of the callback itself can act as that indicator.

Might be nice to validate embedded data when available, even without the callback, given the callback wouldn't be needed.  (VSCode doesn't care though, current plan is it will always supply the callback).

lexaknyazev:
Since callback isn't needed for glb/base64 data, devs may omit it intentionally in certain setups, yet expect full validation. This means that the current approach (using callback's presence as an indicator) isn't correct anyway.

For some models, validation of binary data may be time-consuming process. What if an asset contains 100 of megs of geometry and someone is trying to edit material's factors? VS Code would repeatedly invoke full validation which may lead to poor performance and/or resources usage.

donmccurdy:
An explicit flag would be nice I think, yeah. Should be separate from the callback, since the callback isn’t relevant to embedded data anyway. Maybe something like `ignoreBufferResources`? If that is set to true, should i expect the `externalResourceCallback` to never be invoked, or still used for images?

(I haven’t got my viewer set up to provide external files to validation yet; there are some edge cases for handling drag-and-drop vs ZIP files...)

lexaknyazev:
I think these are separate questions here:
* Should we toggle all binary data fetching or per resource type (all externals, external images, or external buffers)? This may remain as it is: no callback — no externals.
* Should we toggle potentially slow procedures (reading accessors data) regardless of resources storage?

donmccurdy:
> Should we toggle all binary data fetching or per resource type (all externals, external images, or external buffers)? This may remain as it is: no callback — no externals.

As-is seems OK to me.

> Should we toggle potentially slow procedures (reading accessors data) regardless of resources storage?

A way to disable that would be nice I think, yes.

PR #61, opened 2018-01-18
Add note about embedded buffers and externalResourceFunction

donmccurdy:
I noticed while adding resource validation to https://gltf-viewer.donmccurdy.com/ (it's working great, thanks!) that omitting `externalResourceFunction` not only disables validation of external resources, but also of embedded buffers.

Probably this should either be fixed or noted in the docs; this PR does the latter.


lexaknyazev:
There was a discussion on this in #52. Could you check that `options.validateAccessorData` flag doesn't work when `externalResourceFunction` is omitted?

donmccurdy:
Oops, thanks, needed to update to latest version. Will check that later today but there's probably no issue. :)

PR #94, opened 2018-07-25
2.0.0-dev.2.5

emackey:
Allow some newcomers into the list.  These are already added to the main glTF repo.


emackey:
Not sure if it's worth a release just for this, or if you're planning a new dev branch sometime before SIGGRAPH.  Would like to get the new prefixes in before SIGGRAPH though if it's not too much trouble.

lexaknyazev:
A new dev branch is planned for lots of skinning checks. Not sure when exactly it will happen. 
We can issue a new release before SIGGRAPH in either case, so let's wait couple weeks.

emackey:
This is now published to npm.

PR #122, opened 2019-10-10
Only treat first buffer as .glb-embedded

### LINKED ISSUE #121: Should all buffers with empty uri be treated as GLB buffers?
I'm not sure if this is a validator issue or I'm reading the spec wrong.

Following https://github.com/KhronosGroup/glTF/issues/1680, I started to implement support for producing files alongside what was suggested in the issue - emit "dummy" buffer with no data, with the MESHOPT_compression extension overriding that.

I'm encountering an issue where the validator thinks *both* buffers are stored in the .glb file with the following header:

```
"buffers":[{"byteLength":394128},{"byteLength":1166032}],
```

are stored in .glb file:

```
        "resources": [
            {
                "pointer": "/buffers/0",
                "mimeType": "application/gltf-buffer",
                "storage": "glb",
                "byteLength": 394128
            },
            {
                "pointer": "/buffers/1",
                "mimeType": "application/gltf-buffer",
                "storage": "glb",
                "byteLength": 394128
            }
        ],
```

and getting the following errors as a result:

```
            {
                "code": "BUFFER_NON_FIRST_GLB",
                "message": "Buffer referring to GLB binary chunk must be the first.",
                "severity": 0,
                "pointer": "/buffers/1"
            },
            {
                "code": "BUFFER_EXTERNAL_BYTELENGTH_MISMATCH",
                "message": "Actual data length 394128 is less than the declared buffer byteLength 1166032.",
                "severity": 0,
                "pointer": "/buffers/1"
            }
```

Is this expected behavior, or is the validator not doing the right thing here?

Attaching the example file fwiw:
[BrainStem.glb.zip](https://github.com/KhronosGroup/glTF-Validator/files/3707143/BrainStem.glb.zip)

zeux:
Instead of treating all buffers with non-existent URI as being embedded
into GLB file, we only treat the first buffer as embedded - we now emit
a validation error if the first buffer in GLB file has a URI though.

Fixes #121.

### DIFF
--- a/lib/src/data_access/resources_loader.dart
+++ b/lib/src/data_access/resources_loader.dart
@@ -104,20 +104,21 @@ class ResourcesLoader {
             // Data URI
             info.storage = _Storage.DataUri;
             return buffer.data;
-          } else if (context.isGlb && !buffer.hasUri) {
-            // GLB Buffer
-            info.storage = _Storage.GLB;
-            final data = externalBytesFetch();
-            if (context.validate) {
-              if (i != 0) {
+          } else if (context.isGlb && i == 0) {
+            if (buffer.hasUri) {
+              if (context.validate) {
                 context.addIssue(LinkError.bufferNonFirstGlb);
               }
-
+              return null;
+            } else {
+              // GLB Buffer
+              info.storage = _Storage.GLB;
+              final data = externalBytesFetch();
               if (data == null) {
                 context.addIssue(LinkError.bufferMissingGlbData);
               }
+              return data;
             }
-            return data;
           }
         }
         return null;
--- a/lib/src/data_access/resources_loader.dart
+++ b/lib/src/data_access/resources_loader.dart
@@ -104,16 +104,11 @@ class ResourcesLoader {
             // Data URI
             info.storage = _Storage.DataUri;
             return buffer.data;
-          } else if (context.isGlb && i == 0) {
-            if (buffer.hasUri) {
-              if (context.validate) {
-                context.addIssue(LinkError.bufferNonFirstGlb);
-              }
-              return null;
-            } else {
-              // GLB Buffer
-              info.storage = _Storage.GLB;
-              final data = externalBytesFetch();
+          } else if (context.isGlb && !buffer.hasUri && i == 0) {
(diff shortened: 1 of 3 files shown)


zeux:
This change breaks "00_load_buffers_test.dart: Load Buffers Non first buffer refers to GLB" but I'm not sure this test is valid?

The spec says:

> glTF asset could use GLB file container to pack all resources into one file. glTF Buffer referring to GLB-stored BIN chunk, must have buffer.uri property undefined, and it must be the first element of buffers array; byte length of BIN chunk could be up to 3 bytes bigger than JSON-defined 
buffer.byteLength to satisfy GLB padding requirements.

I think this says "the first buffer in a .glb file that has buffer.uri = null needs to refer to BIN chunk" - ok, that's what my change is testing.
I think this does *not* say "the first buffer in a .glb file *must* refer to BIN chunk" - what if no buffers refer to BIN chunk but the chunk is still there? What if no buffers refer to BIN chunk, buffer 1 happens to have uri = null, but the BIN chunk is still there?

It seems to me that while it's not obvious immediately, non_first_buffer.glb is actually valid (it has two buffers, one has a URI, one is just empty/dummy) but I'm not sure if my treatment of the spec is accurate.

lexaknyazev:
> I think this does not say "the first buffer in a .glb file must refer to BIN chunk" - what if no buffers refer to BIN chunk but the chunk is still there? What if no buffers refer to BIN chunk, buffer 1 happens to have `uri` = null, but the BIN chunk is still there?

These are all good questions that fall into "under-defined" spec territory. Before doing anything here, let's clarify the spec first. Could you please open a new spec issue (referring the original use-case of the proposed extension) that would clearly list all questionable corner cases? Thanks very much.

zeux:
I've updated the code to match the spec clarification, and updated the tests to match. CI fails due to dartanalyzer complaining about the code which seems unrelated...

zeux:
@lexaknyazev this should be ready for review since the spec questions have been cleared up

lexaknyazev:
This issue was fixed internally as a part of `2.0.0-dev.3.0` release.

PR #131, opened 2020-01-18
Version 2.0.0-dev.3.2

lexaknyazev:
### Bugfixes

* Validating an asset with an extension that alters allowed enum values (like MIME types for images or vertex formats for accessors) no longer affects subsequent validation runs (introduced in `2.0.0-dev.3.0`).

---

This bug could have been observed only _after_ validating assets with `EXT_texture_webp` or `KHR_mesh_quantization` extensions.

/cc @zeux

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 * Validating an asset with an extension that alters allowed enum values (like MIME types for images or vertex formats for accessors) no longer affects subsequent validation runs (introduced in `2.0.0-dev.3.0`).
 
+* Special characters (`~` and `/`) in JSON Pointers (RFC 6901) are now properly escaped.
+
 ## 2.0.0-dev.3.1
 
 ### New Features
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2.0.0-dev.3.2
+
+### Bugfixes
+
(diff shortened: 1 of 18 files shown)


lexaknyazev:
Also, special characters (`~` and `/`) in JSON Pointers (RFC 6901) are now properly escaped. This can be seen with assets like
```json
{
    "asset": {
        "version": "2.0"
    },
    "asset/customProperty": true
}
```
According to RFC 6901, the correct pointer to the custom property is `/asset~1customProperty`.

emackey:
[MoreBustedBox.zip](https://github.com/KhronosGroup/glTF-Validator/files/4087721/MoreBustedBox.zip)

Another crazy test asset.  The validator does fine with this on this branch.  VSCode's JSON validation complains of a "repeated attribute" that the validator doesn't report (but I'm not sure if the validator is intended to report basic JSON failures).

lexaknyazev:
> "repeated attribute" that the validator doesn't report

That's expected. The only way to detect this is to perform JSON parsing manually instead of relying on platform's `JSON.parse`. ECMAScript definition is:
> In the case where there are duplicate name Strings within an object, lexically preceding values for the same key shall be overwritten.

emackey:
`2.0.0-dev.3.2` published to npm.

/cc @donmccurdy @bghgary

lexaknyazev:
@donmccurdy 
Please update `three-gltf-viewer` to the latest validator build.

donmccurdy:
Updated. ✅

Issue #134, opened 2020-02-04
Validator as a nuget package.


vpenades:
Now that there's a gltf-validator available as a command line tool, I've been using it internally, wrapped as a convenient C# library that exposes an API to validate gltf files.

I was thinking in publishing the nuget myself, but given that 99% of the code is the gltf-validator itself, I was thinking that maybe you guys would want to have ownership and control of it.

I would be willing to do a pull request with the library prototype, something like gltf2validator.

lexaknyazev:
Sounds interesting. I'm not very familiar with nuget ecosystem though. AFAIK, Microsoft glTF-SDK is no longer published there, for some reason.

@bghgary 
What do you think?

bghgary:
If it's wrapped in a C# wrapper, it would make a lot of sense to publish it as a NuGet package. Khronos already has a NuGet account that we can use.

vpenades:
So, there's an agreement in a pull request?

lexaknyazev:
I think a `nuget` folder would be a proper location for the C# project and its sources (just like `node` contains pieces for building a package for `npm`).

A few comments about overall integration.
- Precompiled executable should never be checked-in into the source tree because it depends on the used Dart SDK version, CPU architecture, and operating system. Instead, it should be built on demand.
  - This project uses `tool/grind.dat` script to manage such tasks. We should have a new build task that compiles the native exe and then builds the nuget package into `build/nuget`.
- Assuming that there's nothing Windows-specific, C# project should target .NET Standard to be compatible with multiple .NET implementations.

There's no need to worry about this stuff for now. Please feel free to open a PR with some C# sources and we'll start from there.

vpenades:
Well... since I am doing glTF validation in different c# projects, and copy/pasting the wrapping code has become a hassle, and given that the plan of contributing it here was unsuccessful, I decided to create a separate repository so I can maintain and publish a nuget package that wraps gltf-validator.

[.Net wrapper for gltf-validator](https://github.com/vpenades/GltfValidator)

I would have preferred this project to be handled under khronos umbrella, but I really needed this to be pushed ASAP.

So sorry if this has caused any inconvenience.

emackey:
Hey @vpenades I'm glad to see this published!  May we link to your NuGet package in the README here?

vpenades:
@emackey sure, no problem with that.

PR #136, opened 2020-02-04
Nuget package, first commit

vpenades:
As discussed in #134, In this pull request I dropped the basic idea of a nuget package.

Its basic usage would be this:

```c#
var report = glTF2Validator.ValidationContext.ValidateFile("somefile.gltf");

if (report.HasErrors) throw new Exception(report.RawReport);
```

In order to compile the project, it is required to compile the CLI tools beforehand, since they're embedded in the nuget package.

I have only tested the Windows path; theoretically it should work on Linux, but someone will need to do the testing.

I haven't defined anything for the semantic versioning. Ideally, the nuget package should have the same version ID as the gltf-validator itself.

### DIFF
--- a/nuget/ValidationResult.cs
+++ b/nuget/ValidationResult.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Text;
+using System.Text.Json;
+using System.Text.Json.Serialization;
 
 namespace glTF2Validator
 {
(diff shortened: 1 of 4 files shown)


lexaknyazev:
Looks promising!

I'm concerned about relying API integration on text-based validation output. Its format slightly changes between releases and it does not include all the information that's available in a generated JSON report.

I could add a new command-line key that would print the JSON report to the `stdout` instead of writing it to a file (the latter is certainly not expected from API integration).

vpenades:
> I could add a new command-line key that would print the JSON report to the `stdout` instead of writing it to a file (the latter is certainly not expected from API integration).

Certainly, having the json available to the stdout would greatly help.

lexaknyazev:
CLI option has been added. Please rebase this PR onto `dev` branch and retarget it there.

vpenades:
For some reason, it does not allow me to choose `dev `branch.

lexaknyazev:
Just switched manually. Anyway, all files here are new, so it's safe.

nuget/ValidationResult.cs
```diff
@@ -9,64 +11,49 @@ namespace glTF2Validator
...
-
-            var www = new List<string>();
-            var eee = new List<string>();
-
-            foreach (var l in lines)
+            var options = new JsonSerializerOptions
             {
-                if (l == "\tWarnings:") { status = 1; continue; }
-                if (l == "\tErrors:") { status = 2; continue; }
+                AllowTrailingCommas = true
```
lexaknyazev: Why is this needed? If the report contains trailing commas, it's a bug.

vpenades: Ah, I'll remove it if it's not needed.

vpenades:
@lexaknyazev sorry, I've been working in a different machine, and I didn't notice the author email being used was very old. I've been trying to fix it but it's taking me too much time; so I'm going to remove this pull request and create a new one....

sorry for the inconvenience

lexaknyazev:
No worries! 
FYI, it's possible to force-push a branch that is linked with a PR - GitHub handles that fine.

PR #137, opened 2020-02-06
Nuget package

vpenades:
As discussed in #134, In this pull request I dropped the basic idea of a nuget package.

Its basic usage would be this:

```c#
var report = glTF2Validator.ValidationContext.ValidateFile("somefile.gltf");

if (report.HasErrors) throw new Exception(report.RawReport);
```

In order to compile the project, it is required to compile the CLI tools beforehand, since they're embedded in the nuget package.

I have only tested the Windows path; theoretically it should work on Linux, but someone will need to do the testing.

I haven't defined anything for the semantic versioning. Ideally, the nuget package should have the same version ID as the gltf-validator itself.

### DIFF
--- a/nuget/ValidationContext.cs
+++ b/nuget/ValidationContext.cs
@@ -1,0 +1,78 @@
+﻿using System;
+using System.Runtime.InteropServices;
+
+namespace glTF2Validator
+{
+    /// <summary>
+    /// Represents the context used to validate files
+    /// </summary>
+    public class ValidationContext
+    {
+        /// <summary>
+        /// Path to the gltf-validator command line executable (platform specific).
+        /// </summary>
+        public static string ValidatorExePath { get; set; }
+
+        /// <summary>
+        /// Timeout to cancel the validation operation.
+        /// </summary>
+        public static TimeSpan TimeOut { get; set; } = TimeSpan.FromSeconds(10);
+        
+        static ValidationContext()
+        {
+            SetDefaultPath();
+        }
+
+        
+        private static void SetDefaultPath()
+        {
+            var assemblyDir = System.IO.Path.GetDirectoryName(typeof(ValidationContext).Assembly.Location);
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                if (RuntimeInformation.OSArchitecture != Architecture.X64) return;
+                ValidatorExePath = System.IO.Path.Combine(assemblyDir, "gltf_validator.exe");
+            }
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+            {
+                ValidatorExePath = System.IO.Path.Combine(assemblyDir, "gltf_validator");
+            }
+        }
+        
+        /// <summary>
+        /// Validates a glTF file.
+        /// </summary>
+        /// <param name="gltfFilePath">Path to the glTF file.</param>
+        /// <returns>A report.</returns>
+        public static ValidationReport ValidateFile(string gltfFilePath)
+        {
+            if (!System.IO.File.Exists(gltfFilePath)) throw new ArgumentException($"File {gltfFilePath} not found.", nameof(gltfFilePath));
+
+            if (string.IsNullOrWhiteSpace(ValidatorExePath)) return null;
+
+            if (!System.IO.File.Exists(ValidatorExePath)) throw new System.IO.FileNotFoundException(ValidatorExePath);
+
+            if (!System.IO.Path.IsPathRooted(gltfFilePath)) gltfFilePath = System.IO.Path.GetFullPath(gltfFilePath);
+
+            var psi = new System.Diagnostics.ProcessStartInfo(ValidatorExePath);
+            psi.Arguments = $"-p -r -a -o \"{gltfFilePath}\"";
+            psi.UseShellExecute = false;
+            psi.RedirectStandardOutput = true;
+
+            using (var p = System.Diagnostics.Process.Start(psi))
+            {
+                if (!p.WaitForExit(TimeOut.Milliseconds))
+                {
+                    try { p.Kill(); } catch { }
+                }
+
+                var jsonReport = p.StandardOutput.ReadToEnd();
+
+                if (string.IsNullOrWhiteSpace(jsonReport)) return null;
+
+                return ValidationReport.Parse(jsonReport);
+            }
+        }
+    }
+}
--- a/nuget/ValidationContext.cs
+++ b/nuget/ValidationContext.cs
@@ -1,77 +1,242 @@
-﻿using System;
-using System.Runtime.InteropServices;
+﻿/*
+ * # Copyright (c) 2020 The Khronos Group Inc.
+ * #
+ * # Licensed under the Apache License, Version 2.0 (the "License");
+ * # you may not use this file except in compliance with the License.
+ * # You may obtain a copy of the License at
+ * #
+ * #     http://www.apache.org/licenses/LICENSE-2.0
+ * #
+ * # Unless required by applicable law or agreed to in writing, software
+ * # distributed under the License is distributed on an "AS IS" BASIS,
+ * # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * # See the License for the specific language governing permissions and
+ * # limitations under the License.
+ */
 
-namespace glTF2Validator
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+using System.Runtime.InteropServices;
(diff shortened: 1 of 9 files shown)


vpenades:
@lexaknyazev so far, what's left is:

- Amend the csproj with copyright/license/author stuff, and I guess setting the output path for the package.
- Setup an automated build mechanism that builds this project after the CLI tools have been built, that also passes the semantic version.

I believe it's better to leave these last steps to you, since I don't know the details on how your CI is defined.

lexaknyazev:
Sounds great! I'll review and finish integration early next week.

lexaknyazev:
It seems that NuGet lacks a straightforward way to provide multiple native binaries with a single .NET wrapper, so a few changes were made to the package.

Instead of copying validator's executable into the output folder of every project that depends on the package, now the wrapper library unpacks proper executable from embedded resources on demand. This should make it possible to use the same DLL on all platforms. 

Among other changes:
- Class used for deserialization does not contain any functions anymore.
- More exceptions are caught in advance with more sound messages.
- Async API is provided so that a single `ValidationContext` instance can handle multiple validation requests at the same time. See examples in the README for more info.

nuget/ValidationContext.cs
```diff
@@ -0,0 +1,243 @@
...
+
+            _embeddedValidatorPath = Path.Combine(Path.GetTempPath(), $"{Path.GetRandomFileName()}_{exeName}");
+
+            using (var fileStream = File.OpenWrite(_embeddedValidatorPath))
+            {
+                resourceStream.CopyTo(fileStream);
+            }
+
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Linux) || RuntimeInformation.IsOSPlatform(OSPlatform.OSX))
+                Process.Start("chmod", $"+x {_embeddedValidatorPath}");
```
bghgary:
```suggestion
                Process.Start("chmod", $"+x \"{_embeddedValidatorPath}\"");
```

nuget/ValidationContext.cs
```diff
@@ -0,0 +1,243 @@
...
+        /// <exception cref="InvalidOperationException">
+        ///     Thrown when the <c>ValidationContext</c> was created without embedded
+        ///     validator but custom one was not provided.
+        /// </exception>
+        /// <exception cref="TimeoutException">Thrown when validation takes longer than allowed time.</exception>
+        /// <exception cref="ApplicationException">Thrown when the validator encounters unhandled runtime error.</exception>
+        public ValidationReport Validate(string path, string customValidatorPath = "", bool validateResources = true,
+            TimeSpan timeout = default)
+        {
+            return ValidateAsync(path, customValidatorPath, validateResources, timeout).Result;
```
bghgary: Does this work? I think you have to wait for the task to complete.

nuget/ValidationContext.cs
```diff
@@ -0,0 +1,243 @@
...
+        /// <returns>A <c>ValueTask</c> completing with <c>ValidationReport</c> instance.</returns>
+        /// <exception cref="ArgumentException">Thrown when either asset or custom glTF-Validator path is incorrect.</exception>
+        /// <exception cref="FileNotFoundException">Thrown when glTF asset or custom glTF-Validator is not found.</exception>
+        /// <exception cref="InvalidOperationException">
+        ///     Thrown when the <c>ValidationContext</c> was created without embedded
+        ///     validator but custom one was not provided.
+        /// </exception>
+        /// <exception cref="TimeoutException">Thrown when validation takes longer than allowed time.</exception>
+        /// <exception cref="ApplicationException">Thrown when the validator encounters unhandled runtime error.</exception>
+        public async ValueTask<ValidationReport> ValidateAsync(string path, string customValidatorPath = "",
```
bghgary: I'm not sure why ValueTask is returned here. It probably makes more sense to return Task<ValidationReport> and call Wait on it in the synchronous version.

vpenades:
> I've never seen anyone extract resources into an executable in a temp folder and then run it. I'm not sure what the security implications are.

My original idea was that the command line exe is included as part of your application, not copied to tmp...

The only other alternative would be to compile the cli as a DLL and not as an EXE, and call it with p/invoke, but I have no idea how to do that or if it's even possible for Dart to compile a native dll.

Certainly it's not common, but it's not that uncommon, I've seen a few nugets around that have no other choice that include an exe that does the stuff.

I guess it's fine with due warning. Personally, I want to use it in several unit tests I have around.
My point is that any help you give to developers to validate the exported models will help improve the whole ecosystem as a whole.

nuget/ValidationContext.cs
lexaknyazev: AFAIU, calling `.Result` on `ValueTask` should handle that internally. Maybe need to double-check.

lexaknyazev:
>  command line exe is included as part of your application

I am quite concerned about force-adding a new `.exe` to someone's app folder. Nevertheless, it's still possible to build the .NET wrapper without embedded resources (just leave `compiled` folder empty) and always set a path to app-provided validator.

> I've never seen anyone extract resources into an executable in a temp folder and then run it. I'm not sure what the security implications are.

I think this pattern is quite common among software installers or wrappers for "portable" apps. It seems more secure than running executable that is already present on the disk and could have been replaced.

vpenades:
> I am quite concerned about force-adding a new `.exe` to someone's app folder. Nevertheless, it's still possible to build the .NET wrapper without embedded resources (just leave `compiled` folder empty) and always set a path to app-provided validator.

Just for the sake of reference, I did a small experiment; I went to my nuget local cache directory (typically located in `Users\[You]\.nuget\packages` and I've run a search of all the *.exe files...  and there's quite a few embedded executable CLI tools inside packages

To name some packages that embed executables:
- `microsoft.codecoverage`
- `microsoft.servicefabric.actors`
- `microsoft.testplatform.testhost`
- `xunit.runner.console`
- `geckofx60.32`
- `eo.webbrowser`
- `wix`
- `mono.apitools`
- `mono.webassembly.build`
- `pdfbox`
- 'diskutils`
- 'fbxsharp`

Notice that all these are "normal" packages, not dotnet tool packages.

So I don't think it's that uncommon to embed tools inside packages; and I understand the security concerns, but it's the gltf-validator tool you're including, not some obscure third party exe.

So my opinion is that, for the sake of ease of use, I would preffer to keep the exe files within the package. This will also be specially helpful when automatically updating the nuget package, which will automatically bring new versions of the actual gltf-validator.... otherwise, you need to update the nuget package, and manually update the exe files.

nuget/ValidationContext.cs
bghgary: In the Remarks of the [documentation](https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.valuetask-1?view=netcore-3.1#remarks), it states:

> The following operations should never be performed on a ValueTask<TResult> instance:
> ...
> Using .Result ... when the operation hasn't yet completed...
> ...

bghgary: Also
> consumers may not read Result until the instance has completed

lexaknyazev: Oops, it's `Task.Result` that waits until completion, not `ValueTask.Result`.
> Accessing the property's get accessor blocks the calling thread until the asynchronous operation is complete; it is equivalent to calling the `Wait` method.

vpenades:
As an additional note about whether embedding the Exe files inside the nuget packages (which is my preferred choice), the final decision is up to you, but if you choose to not embed the exe in the package, requiring the developers to manually update the exe files along with the nuget package....

I would propose that the nuget code to check the version of the gltf-validator tool to be in line with the one of the nuget package.  My fear is that people unaware of this detail may update the nuget package and keep an old version of the gltf-validator around.

vpenades:
What's the status of this PR? I guess everybody's in containment... ?

lexaknyazev:
Sorry, the PR got auto-closed. If it's still relevant please reopen.

vpenades:
Well... it all depends on whether you're still interested in officially supporting a release of glTF validator as a nuget.

Right now I'm using glTF validator for unit testing directly (using essentially the same calls and logic that would be added to the nuget project). Now that you've published a new release, I'll manually update my project... but it could have been much easier if I would had so simply update a nuget package. This is not a big deal given the current rate of releases (2, 3 per year), but if rate increases is a hassle.

The other reasons for having a nuget package is for other users using c# to easily add additonal validations to their projects, where these scenarios might apply:
- importing models from unknown sources, and the importer's validation is not as robust as glTF validation.
- Testing while developing importers/exporters.
- Enabling strict mode on exporters. (this could be useful for max and blender exporters, to systematically call glTF validator on export)

So again, it all boils down to whether you might want to follow that path or not...

PR #142, opened 2020-04-13
Remove require assignment from node preamble

### LINKED ISSUE #139: Warning when running within Angular
When running within angular, we receive the following error:

./node_modules/gltf-validator/gltf_validator.dart.js 1:243-250
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

robertlong:
Fixes #139 

Note that this breaks require which I don't think is an issue now but would be in the future if there were external node dependencies.

### DIFF
--- a/tool/grind.dart
+++ b/tool/grind.dart
@@ -133,11 +133,19 @@ void _npmBuild({bool release = true}) {
   // TODO: remove this patch after upstream update
   {
     final file = File(p.join(_nodeTarget, 'gltf_validator.dart.js'));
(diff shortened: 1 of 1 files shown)


emackey:
Hi @robertlong, sorry this got overlooked.  There's a new fix for this as part of #148.

Hubs is very cool BTW, I hope that project keeps going in these COVID times.

lexaknyazev:
This fix was incorporated into `2.0.0-dev.3.3`.

Issue #146, opened 2020-08-13
Multiple Extensions warning


emackey:
With the upcoming PBR Next extensions, multiple extensions on a single material object are expected and encouraged.  The validator should not issue a warning for this finding.

Two exceptions: The `KHR_materials_unlit` extension, and the `KHR_materials_pbrSpecularGlossiness` extension, may not co-exist with any of the upcoming PBR Next extensions.  Ideally, the validator should detect that and report a full error, as the specifications have been updated to exclude those two explicitly.

All of the other planned KHR material extensions may co-exist within a single material.  We could issue a warning if a single material calls for more than some number of textures at once (5 or 6?) as low-end implementations could run out of texture samplers (taking environment and LUT into account).

lexaknyazev:
Fixed in `2.0.0-dev.3.3`.

aaronfranke:
This warning is showing up when `KHR_materials_unlit` is used with `VRMC_materials_mtoon`:

<img width="856" height="269" alt="Image" src="https://github.com/user-attachments/assets/43634ffb-118c-44ec-8541-677235f23e3e" />

The `VRMC_materials_mtoon` extension is designed to be used together with `KHR_materials_unlit`, since the intention is that when `VRMC_materials_mtoon` is not supported, the `KHR_materials_unlit` extension serves as a fallback.

Issue #145, opened 2020-08-06
No explicit warning when using unlit and pbrSpecularGlossiness


UX3D-haertl:
Currently, there is only a general warning if multiple extensions are used in one object.
But in the unlit spec this is stated:

> **Implementation Note:** When `KHR_materials_unlit` is included with another
> extension specifying a shading model on the same material, the result is
> undefined.

Therefore, if unlit and pbrSpecularGlossiness exist in a material a more explicit warning about the undefined state might be needed.

emackey:
Yes.  Almost all of the upcoming "PBR Next" set of extensions are expected to play well with each other, and specifically exclude both `KHR_materials_unlit` and `KHR_materials_pbrSpecularGLossiness` from being in the same material.

lexaknyazev:
Fixed in `2.0.0-dev.3.3`.

Issue #141, opened 2020-04-07
Validator gets stuck in endless loop


Silverlan:
I've tried validating the attached glTF file, but the validator just gets stuck at "Validating..." endlessly and never seems to progress. The file does not contain any meshes, only a skeleton and an animation.
I've only tried the web-version of the validator (https://github.khronos.org/glTF-Validator/).

[soldier.zip](https://github.com/KhronosGroup/glTF-Validator/files/4446149/soldier.zip)

lexaknyazev:
Confirmed.

javagl:
Just in case that it *might* be helpful: The same appeared for a model that is simpler than the attached "soldier" model:

<sup>(Note that the following model does not have a buffer URI, it was only extracted from a GLB, but this seems to be unrelated to the issue)</sup>

```
{
  "scene":0,
  "scenes":[
    {
      "name":"Scene",
      "nodes":[
        0
      ]
    }
  ],
  "nodes":[
    {
      "mesh":0,
      "name":"Mesh"
    }
  ],
  "meshes":[
    {
      "name":"Mesh",
      "primitives":[
        {
          "indices":0,
          "attributes":{
            "POSITION":1,
            "TEXCOORD_0":2,
            "COLOR_0":3,
            "NORMAL":4
          }
        }
      ]
    }
  ],
  "accessors":[
    {
      "bufferView":0,
      "componentType":5125,
      "count":12,
      "max":[
        4
      ],
      "min":[
        0
      ],
      "type":"SCALAR"
    },
    {
      "bufferView":1,
      "componentType":5126,
      "count":5,
      "max":[
        1.0,
        1.0,
        2.0
      ],
      "min":[
        -1.0,
        -1.0,
        -0.0
      ],
      "type":"VEC3"
    },
    {
      "bufferView":2,
      "componentType":5126,
      "count":5,
      "max":[
        1.0,
        1.0
      ],
      "min":[
        0.0,
        0.0
      ],
      "type":"VEC2"
    },
    {
      "bufferView":3,
      "componentType":5126,
      "count":5,
      "max":[
        1.0,
        1.0,
        1.0
      ],
      "min":[
        0.0,
        0.0,
        0.0
      ],
      "type":"VEC3"
    },
    {
      "bufferView":4,
      "componentType":5126,
      "count":5,
      "max":[
        0.5773503184318543,
        0.5773503184318543,
        1.0
      ],
      "min":[
        -0.5773503184318543,
[...]

Issue #140, opened 2020-03-08
Camera perspective yfov has no upper limit.


vpenades:
The test/base/data/camera/extra_type.gltf defines camera perspective YFOV with value 10.

The official specification states that the value must be over 0, and although it does not explicitly state the upper limit, it states that the value is defined as radians, which sets an implicit upper limit.

Also, the spec says the value must be used in formulas using the form `Tan(0.5*yfov)`  which implicitly sets the upper limit _below π_.

I guess the spec does not define the upper limit because it's a non rational number?  but math libraries certainly define it...

So I guess at the very least the gltf_validator should take this into account.

lexaknyazev:
Opened a spec issue. Will add validation warning anyway.

lexaknyazev:
Fixed in `2.0.0-dev.3.3`.

PR #148, opened 2020-09-25
Version 2.0.0-dev.3.3

lexaknyazev:
### New Features

* Added support for new PBR extensions: `KHR_materials_clearcoat`, `KHR_materials_transmission`, and `KHR_materials_sheen`.

* Added `CAMERA_YFOV_GEQUAL_PI` validation warning (#140).

* Added new `SKIN_IBM_ACCESSOR_WITH_BYTESTRIDE` error for IBM accessors using buffer views with defined `byteStride`.

* Added new `IMAGE_BUFFER_VIEW_WITH_BYTESTRIDE` error for images using buffer views with defined `byteStride`.

* Added new vendor prefixes: `ANIMECH`, `GRIFFEL`, `MAXAR`, `MPEG`, `PANDA3D`, `PTC`, `SEIN`, `SPECTRUM`, `TRYON`, and `VRMC`.

### Changes

* Modified `MULTIPLE_EXTENSIONS` warning. It is now reported only for explicitly standalone extensions (#145 and #146).

* Added missing check for morph target accessors overriding buffer view usage.

### Bugfixes

* Fixed an endless loop on assets with a node loop within a scene (#141).

* Fixed building all targets at once (`pub run grinder` with no args).

* Fixed missing JS API error messages when the validator is compiled in debug mode.

* Web drag-n-drop tool now shows a correct message when no glTF assets were provided.

### Integration updates

  * npm tool no longer re-defines `require` (#139).

  * npm tool no longer requires `options.externalResourceFunction` for GLB or assets with embedded data (#138).

  * `options.validateAccessorData` npm tool flag is removed, accessors' data validation is always enabled.

  * npm tool now reports `IO_ERROR` for external resources when `options.externalResourceFunction` is not provided.

  * CLI tool now supports `--stdout` flag for writing a single validation report to STDOUT instead of a new file.

  * CLI tool validates binary data by default (#138).


donmccurdy:
Unfortunately it'll be 1-2 weeks before I'm able to review this — if others have time to review in the meantime, please feel free to merge this without me.

> Should we fail in a case when an npm version is called without options.externalResourceFunction but the asset uses external files and their validation was requested with options.validateAccessorsData: true?

Because `validateAccessorsData` defaults to true and `externalResourceFunction` defaults to undefined, it seems unfortunate that validation would always fail with its default options... Maybe providing `externalResourceFunction` should simply cause accessor data to be validated, and the `validateAccessorsData` option can be removed?

emackey:
> Maybe providing `externalResourceFunction` should simply cause accessor data to be validated, and the `validateAccessorsData` option can be removed?

I like this idea.  Basically if the external data can be validated (either because it's GLB or because a function was supplied), then there's no reason to skip validation of it.

If such validation doesn't happen, perhaps there could even be a warning, that the model wasn't completely submitted to the validator?

lexaknyazev:
>  then there's no reason to skip validation of it

I think there were some reasons initially. The main was the assumption that binary data validation could be too slow but it looks like even with all the extra checks added in the previous update it's still fast enough.

> If such validation doesn't happen, perhaps there could even be a warning, that the model wasn't completely submitted to the validator?

How should such warning be reported? Since it doesn't apply to the asset itself, we'd need to adjust the report schema to have an extra flag, like it has `truncated` for reports with too many messages. Web version would have an extra visual warning that not all external resources were submitted (as some users drag-n-drop only the JSON file).

emackey:
> How should such warning be reported?

I'm flexible on this.  We do have some warnings that are not asset-related, primarily I'm thinking of the "extension not supported by the validator" warning.  But yes, this could be a case for a new "truncated" type of warning.

Basically my thinking is, if a user has gone to the trouble to run the validator, they should be warned if it wasn't a completely successful validation of all parts of the glTF, including the external resources.  I'm open to ideas though.

lexaknyazev:
Fair point. We could attribute the lack of `externalResourceFunction` to resources that actually need it rather than to the report in general.

emackey:
Ah, we could apply this warning using JSON pointers for each of the external resource(s) affected, that sounds good to me.

emackey:
@donmccurdy Are you OK with where this conversation ended?

ISSUES.md
```diff
@@ -106,6 +108,7 @@
 |NODE_SKIN_WITH_NON_SKINNED_MESH|Node has skin defined, but mesh has no joints data.|Error|
 |NODE_WEIGHTS_INVALID|The length of weights array (`%1`) does not match the number of morph targets (`%2`).|Error|
 |SCENE_NON_ROOT_NODE|Node `%1` is not a root node.|Error|
+|SKIN_IBM_ACCESSOR_WITH_BYTESTRIDE|bufferView.byteStride must not be defined for buffer views used by IBM accessors.|Error|
```
emackey, line 111: In the long description here, can we spell out IBM to `inverseBindMatrices`?

emackey (approved):
Small suggestion here, but this all looks great.  I tested some of the new warnings and messages in VSCode.

ISSUES.md
lexaknyazev: Since the error is already attributed to `inverseBindMatrices` property, maybe spell out IBM in the description using separate words?

emackey: Either way is fine.  Reading the `ISSUES.md` list by itself didn't make this clear, as there was no attribution.  But I think it's fine for the reported error to include the name `inverseBindMatrices` even when it is attributed.

lexaknyazev:
@emackey 
Is sheen extension schema ready to be added here?

emackey:
Yes, thanks!

lexaknyazev:
Added `KHR_materials_sheen` support.

donmccurdy:
> Are you OK with where this conversation ended?

A warning on external resources that couldn't be validated sounds good to me, yes. 👍

emackey:
Let's add the UX3D extension prefix here and call this one done. ✅

lexaknyazev:
@emackey 
Let me update the SDK constraints to keep CI green since we're not fully compatible with 2.12 yet.

emackey:
Anything I can do to help this along?  Do we just need a tighter range on the SDK listed in `pubspec.yaml`?

.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: dart
 sudo: false
 dart:
-  - dev
+  - beta
```
emackey, line 4: Can/should we just use dart stable, instead of beta?

web/scripts/validator.dart
```diff
@@ -16,6 +16,8 @@
 
 // ignore_for_file: avoid_print
 
+// @dart=2.9
```
emackey, line 19: Why does this say 2.9 when the range asks for 2.10?

emackey (commented):
Travis has been running for almost 2 hours, is it stuck?

web/scripts/validator.dart
lexaknyazev: The docs suggest to use `2.9` as the language version here. Anyway, I'll remove it in the next release.

.travis.yml
lexaknyazev: It's good to be aware of any forthcoming breaking changes, so I'd leave both (it didn't cause any issues in the past).

emackey:
Published to [npm](https://www.npmjs.com/package/gltf-validator).

lexaknyazev:
[Drag-n-drop tool](https://github.khronos.org/glTF-Validator/) updated.

hypnosnhendricks:
How could I have tracked the development of SKIN_IBM_ACCESSOR_WITH_BYTESTRIDE from when it was a concept to implementation to release? Our GLB assets are currently running into this newly added error & I would like to be able to track progress of any new errors being added to the validation tool before they make it into release. My github skills are basic at best. Thanks!

lexaknyazev:
@hypnosnhendricks 
The easiest way would be to subscribe to the repo updates as described [in the GitHub docs](https://docs.github.com/en/free-pro-team@latest/github/getting-started-with-github/be-social#watching-a-repository).

As for that particular validation message, its omission before this release was a bug because the spec defines `byteStride` only for vertex attributes.

PR #158, opened 2021-04-13
Version 2.0.0-dev.3.4

### LINKED ISSUE #154: Vertex attribute with UNSIGNED_INT (5125) componentType is not caught by the validator
We received a model that has a vertex attribute with the `UNSIGNED_INT (5125)` `componentType` which isn't isn't valid according to the glTF spec but passes the validator. The model does not have any extensions.

> "gltf_detailedDescription": "The datatype of components in the attribute.  All valid values correspond to WebGL enums.  The corresponding typed arrays are `Int8Array`, `Uint8Array`, `Int16Array`, `Uint16Array`, `Uint32Array`, and `Float32Array`, respectively.  **5125 (UNSIGNED_INT) is only allowed when the accessor contains indices**, i.e., the accessor is only referenced by `primitive.indices`.",

I'm still waiting to see if we can share the model but in the meantime here are some snippets from the glTF.

```json
{
  "bufferView":7,
  "componentType":5125,
  "count":2145,
  "type":"SCALAR",
  "min":[4],
  "max":[4]
}
```

```json
"primitives": [
  {
    "attributes": {
      "_MaxHandle": 7,
      "NORMAL": 6,
      "_fbx_rotation": 4,
      "TEXCOORD_0": 5,
      "_fbx_scale": 3,
      "_fbx_translation": 2,
      "POSITION": 1
    },
    "indices": 0,
    "mode": 4,
    "material": 0
  }
]
```

`_MaxHandle` points to the accessor above.

CC https://github.com/CesiumGS/cesium/issues/9270

### LINKED ISSUE #159: Consider lowering the severity of UNSUPPORTED_EXTENSION
I recommend lowering the severity of `UNSUPPORTED_EXTENSION` from "warn" (1) to "info" (2).  The glTF format is very extensible, and we shouldn't be warning users not to extend it just because we're unable to validate those extensions.

VSCode 1.55 has increased the default strength and visual noise of warnings.  For example:

image

Across the title bar, glTF models with vendor extensions have their names turn from white to yellow, and a little number appears next to them.  This treatment is not given to "info" or "hint", just "error" (red) and "warn" (seen here).

Similar treatment is given to the list of files when the file tree is visible.  Files that use vendor extensions are marked in yellow with numbers next to them, while core files can be shown normally in white.

While this app is only one use of the validator, I think it's not unreasonable for any app to be calling out the warnings like this.  But vendor extensions should be treated as natural and harmless, not called out as warnings.

lexaknyazev:
### New Features

* Added support for `KHR_materials_variants` extension.

* Added support for new PBR extensions: `KHR_materials_ior`, `KHR_materials_specular`, and `KHR_materials_volume`.

* Added `MESH_PRIMITIVE_ATTRIBUTES_ACCESSOR_UNSIGNED_INT` validation error, fixes #154.

* Added new vendor prefixes: `ASOBO`, `CITRUS`, `CLO`, `OFT`.

### Bugfixes

* Fixed incomplete JSON pointers for out-of-range array elements.

* Web build is now usable from a local filesystem.

### Integration updates

* Added `exe-archive` and `web-archive` tasks to generate release archives.

* Fixed test names for `KHR_lights_punctual` extension.

* Changed default severity of `UNSUPPORTED_EXTENSION` to Info, fixes #159.


donmccurdy:
One suggestion, otherwise changes look good!

Might be worthwhile to check for the (disallowed) combination of PBR Next extensions with `KHR_materials_pbrSpecularGlossiness` or `KHR_materials_unlit`.

There's also an interaction with volume and transmission...

> The volume extension needs to be combined with an extension that allows light rays to pass through the surface into the volume, e.g. KHR_materials_transmission

... but perhaps no point in testing for that case until the translucency extension is also available.

emackey (approved):
This is looking good so far.  It will be great to have support for specular/ior/volume.

emackey (approved):
Looks awesome, glad to have variants added to the list too.  Tested a couple things in VSCode, seems to work fine.

lexaknyazev:
@donmccurdy 

> Might be worthwhile to check for the (disallowed) combination of PBR Next extensions with `KHR_materials_pbrSpecularGlossiness` or `KHR_materials_unlit`.

Since the previous release, these two extensions get flagged when used with any other material extensions.

emackey:
I hope I didn't jump the gun here, but since this got merged, I published dev3.4 to NPM, built from commit 22a8b17a2b5.

PR #167, opened 2021-06-21
Include docs in archives & update notices

lexaknyazev:
This PR updates 3rd-party licenses and ensures that all required docs are included in the generated archives.

@emackey 
Do we need to re-publish the npm package?


emackey:
@lexaknyazev I suspect it's easier to bump the version number than it is to try to edit a published version on npm.

lexaknyazev:
The already published npm package is immutable. Do you think we should issue a new one just for this change?

lexaknyazev:
I think these are our options:
- Leave the version as-is
- Bump to `2.0.0-dev.3.5`
- Append build metadata like `2.0.0-dev.3.4+1`

For the last two options we may choose to release a new npm package or to wait until the next release.

emackey:
I don't have a strong opinion.  May as well bump to 3.5, just to keep the version numbering simple.

lexaknyazev:
PTAL

emackey (approved):

emackey:
@lexaknyazev Awesome.  Let me know when the GitHub release tag is up, thanks!

lexaknyazev:
The release tag is up. This time, we also ship a precompiled binary for macOS and a copy of the hosted drag-n-drop tool in addition to Windows and Linux builds.

emackey:
This is published to [NPM](https://www.npmjs.com/package/gltf-validator) now.

PR #168, opened 2021-09-21
Add link in README to third-party NuGet package.

### LINKED ISSUE #134: Validator as a nuget package.

emackey:
Fixes #134

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,8 @@ Live drag-n-drop tool: https://github.khronos.org/glTF-Validator
 
 NPM package: https://www.npmjs.com/package/gltf-validator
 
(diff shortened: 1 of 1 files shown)


emackey:
Not sure why CI failed on a README change.

lexaknyazev:
> Not sure why CI failed on a README change.

Because the newer SDK version contains updated code formatting rules.

I'll run a few tests with the nuget package locally and merge the PR afterwards.

PR #179, opened 2022-03-31
Version 2.0.0-dev.3.6

lexaknyazev:
### New Features

* Added support for `KHR_materials_emissive_strength` extension; added `KHR_MATERIALS_EMISSIVE_STRENGTH_ZERO_FACTOR` warning.

* Added `KHR_MATERIALS_VOLUME_NO_TRANSMISSION` warning.

* Added 10 new vendor prefixes.

* Added allowed accessor formats for `TEXCOORD_n` and `COLOR_n` morph target attributes.

* Added `ANIMATION_SAMPLER_ACCESSOR_WITH_BYTESTRIDE` error.

* Added `BUFFER_BYTE_LENGTH_MISMATCH` error.

* Added `BUFFER_VIEW_TARGET_MISSING` hint.

* Added `CAMERA_XMAG_YMAG_NEGATIVE` warning.

* Added `GLB_EMPTY_BIN_CHUNK` info.

* Added `URI_GLB` info.

* Added `UNUSED_MESH_WEIGHTS` and `UNUSED_MESH_TANGENT` infos.

### Changes

* Skipped range-checking for vertex color attributes not defined in the spec.

* Data URIs with content not matching the declared `mediatype` now trigger `INVALID_URI` error.

* Updated `INVALID_IBM_ACCESSOR_COUNT` validation logic and message.

* Updated `MESH_PRIMITIVE_MORPH_TARGET_NO_BASE_ACCESSOR` validation message.

* Fixed JSON-pointers for `IMAGE_MIME_TYPE_INVALID` error.

* Changed `ASSET_MIN_VERSION_GREATER_THAN_VERSION` default severity to Error.

* Changed `CAMERA_XMAG_YMAG_ZERO` default severity to Error.

* Changed `DATA_URI_GLB` default severity to Warning.

* Removed `BUFFER_EMBEDDED_BYTELENGTH_MISMATCH` error.

* Removed `BUFFER_EXTERNAL_BYTELENGTH_MISMATCH` error.

* Removed `INTEGER_WRITTEN_AS_FLOAT` warning.

* Removed `MESH_PRIMITIVE_TANGENT_POINTS` and `MESH_PRIMITIVES_UNEQUAL_JOINTS_COUNT` validation issues.

### Bugfixes

* Fixed false positive `NODE_EMPTY` caused by unknown extensions.

* Fixed assertion on JSON arrays containing numbers that do not fit into double precision.

* Made `BUFFER_DATA_URI_MIME_TYPE_INVALID` validation case-insensitive.

* Tightened JPEG and WebP detection logic.


lib/src/ext/KHR_materials_volume/khr_materials_volume.dart
```diff
@@ -71,6 +71,21 @@ class KhrMaterialsVolume extends GltfProperty {
...
+
+    if (!context.validate) {
+      return;
+    }
+
+    Object o = this;
+    while (o != null) {
+      o = context.owners[o];
+      if (o is Material) {
+        if (!o.extensions.containsKey(KHR_MATERIALS_TRANSMISSION)) {
```
emackey, line 83: Can we allow unknown material extensions to suppress this warning?  There's a draft KHR_materials_translucency that may get renamed to KHR_materials_diffuse_transmission, either of which would dismiss the need for KHR_materials_transmission.  It's also possible (though unlikely) that a vendor-specific extension could allow light through the surface of a volume.

lexaknyazev: The plan is to add transmission-related KHR and EXT extensions here once they become stable. Suppressing the message when a vendor-specific extension is present sounds fine.

lexaknyazev: Looking into this again, we should suppress the warning regardless of the extension prefix when an unknown material extension is present.

javagl (approved):
I had a look at the changes in the individual commits. For some of them, I wondered where the change came from. In some cases, I found the answer in the spec (for example, the change for the "camera xmag and ymag validation" clearly matches the wording of _"...MUST not be 0, SHOULD not be negative"_). But at some point, have to _assume_ that a change like "Removed BUFFER_EXTERNAL_BYTELENGTH_MISMATCH error." has a reason that is beyond what I can understand within a reasonable time frame, so I did not counter-check all changes against the spec systematically.

I tried it out with the latest state, on the sample models, and they seem to pass (except for the `Fox`, which is already tracked in an issue).

lexaknyazev:
@javagl 
Updating the SDK requirement would require using the newer language version that would need another round of internal updates. Keeping it at the current state enables "legacy" mode. I'm going to handle this in the next update - current priority is to publish the updated validation rules.

emackey:
This may just be a fluke, but on Windows OS at the end of `dart run grinder npm` I get this error message:

> FileSystemException: Deletion failed, path = 'build\node\node_modules' (OS Error: The process cannot access the file because it is being used by another process.

emackey:
It looks like `BUFFER_VIEW_TARGET_MISSING` is being produced on almost every primitive attribute coming out of the Blender exporter (as of the version released with Blender 3.0 at least).  I think folks will be unhappy to see this much noise in the validation outputs for such models, when previously these came out clean or mostly clean.

How important is it to let people know about these missing targets?  I think if we're going to leave the message in, we'll need a fix for Blender (and other exporters) and we'll need a major cleanup of the sample models.

emackey:
For comparison, here's `BUFFER_VIEW_TARGET_MISSING` at the current severity 2:

target info screenshot

And again, downgraded to severity 3 hint:

target hint screenshot

The white "..." indicators show where the hints are.  Even so, with enough attributes it's possible to overflow the max number of issues and get a validation error.  But this could be one way to make the message less intrusive in VSCode at least.

lexaknyazev:
> This may just be a fluke, but on Windows OS...

Most likely, this was caused by some AV software that didn't unlock the folder in time.

---

> How important is it to let people know about these missing targets?

The spec says:
> When a buffer view is used by vertex indices or attribute accessors it SHOULD specify `bufferView.target`...

It seems reasonable to reflect such statement in the validator output. Reducing the severity to the minimum seems fine. Additionally, we may consider not counting validation hints into the total messages limit.

---

> we'll need a major cleanup of the sample models

We need that anyway since there are several long-standing issues that clutter the CI output.

emackey (approved):
I tried this with a couple of different sample models in VSCode, and things seem to be working fine.

emackey:
Version 2.0.0-dev.3.6 has been published to [npm](https://www.npmjs.com/package/gltf-validator).

Issue #180, opened 2022-04-26
Web validator crashes (uncaught exception) for specific file


hybridherbst:
While working on the UnityGltf exporter (using the validator to check if resulting files are corrrect) I somehow managed to produce a file that
- works in model-viewer / three.js
- crashes the validator.

I can only assume that it's _really really_ broken, but it would be better if Validator could tell me in which way.

To reproduce, 
- download [validatorException.zip](https://github.com/KhronosGroup/glTF-Validator/files/8566770/validatorException.zip), unpack
- drag "todo (1).glb" into https://github.khronos.org/glTF-Validator/
- Expected: shows a horrible list of issues with this file
- Actual: nothing happens, stuck at "Validating..."

This appears in the browser console:
```
Uncaught Instance of 'minified:d5'
    at Object.c (https://github.khronos.org/glTF-Validator/scripts/validator.dart.js:315:3)
    at i.a8 (https://github.khronos.org/glTF-Validator/scripts/validator.dart.js:8614:9)
    at i.a3 (https://github.khronos.org/glTF-Validator/scripts/validator.dart.js:8624:23)
    at fb.dZ (https://github.khronos.org/glTF-Validator/scripts/validator.dart.js:9869:72)
    at tear_off.<anonymous> (https://github.khronos.org/glTF-Validator/scripts/validator.dart.js:437:65)
    at n3.eI (https://github.khronos.org/glTF-Validator/scripts/validator.dart.js:6250:21)
    at n3.dd (https://github.khronos.org/glTF-Validator/scripts/validator.dart.js:6254:21)
    at eh.b1 (https://github.khronos.org/glTF-Validator/scripts/validator.dart.js:6165:5)
    at di.d5 (https://github.khronos.org/glTF-Validator/scripts/validator.dart.js:6208:9)
    at n1.$0 (https://github.khronos.org/glTF-Validator/scripts/validator.dart.js:6227:3)
```

When dragging the file into https://modelviewer.dev/editor/, it actually has the content that it should have (and _seems_ to be fine), but there the red "ERROR" button actually contains a long list of validation issues.

lexaknyazev:
Thanks for the repro case!

The file surprisingly generates too many (over 16384) validation errors while parsing the GLB container. Apparently, there's no exception handler for this case.

I'll investigate and release an updated build ASAP.

hybridherbst:
Wow, I guess that might get a record for "most issues in a file that still somehow works fine" 😁
[...]

Issue #174, opened 2021-11-02
Extraneous linefeed in ISSUES.md


emackey:
Was looking at ISSUES.md again today, and noticed there's an extra linefeed near the top of the file, in the middle of the description for `INVALID_URI`.

PR #181, opened 2022-04-27
Version 2.0.0-dev.3.7

lexaknyazev:
### New Features

* Added `GLB_EXTRA_DATA` warning. Extra bytes beyond the declared GLB total length are now ignored.

### Bugfixes

* Fixed crash when exceeding the maximum reported issues threshold while parsing GLB structures (#180).

* Fixed erroneous line break in the generated issues table (#174).

### Integration updates

* ZIP-archives generated on Windows can now be opened on other platforms.


tool/grind.dart
```diff
@@ -61,7 +62,7 @@ void issues() {
       } on TypeError catch (_) {
         message = issueType.message(argsWithArray);
       }
-      sb.writeln('|${issueType.code}|$message|'
+      sb.writeln('|${issueType.code}|${message.replaceAll('\n', ' ')}|'
```
javagl, line 65: I wondered about the options of 

- removing the `\n` here
- omitting it at https://github.com/KhronosGroup/glTF-Validator/blob/9ecda14d6d2899d3680182ca93c6aea95a1662c3/lib/src/errors.dart#L248 in the first place

(The "Parser output" info from the `invalidJson` does not have the `\n`...). 
But don't have any basis to say what would be "better". I could imagine that, depending on the output, the `\n` adds clarity or readability.

javagl (approved):
I tried to reproduce issue https://github.com/KhronosGroup/glTF-Validator/issues/180 , but this apparently _only_ happens in the web version? I'm not sure about the translation process from Dart to the minified JavaScript. The "desktop version" (before the fix) seems to work and create
```
{
    "uri": "test/base/data/glb/todo.glb",
    "mimeType": "model/gltf-binary",
    "validatorVersion": "2.0.0-dev.3.5",
    "issues": {
        "numErrors": 22310,
        "numWarnings": 22333,
...
```
with the input file that was given as a test case. (Speaking of test cases: I wondered whether there should be some `glb-with-lots-of-zeros.glb` as "unit test" for that, but maybe the `extra_data.glb` sufficiently covers this)

I don't know what happens in `grind.dart` in all detail, but at least can confirm that with the changes done there, `dart run grinder exe-archive` (still) works.

tool/grind.dart
lexaknyazev: The invalid URI error message consists of two lines, so prepending `\n` makes sense there although the layout of that output format makes sense only with monospace fonts.

In any case, ensuring that the table does not contain extra line breaks (even if there aren't any) is an improvement.

lexaknyazev:
> I tried to reproduce issue https://github.com/KhronosGroup/glTF-Validator/issues/180, but this apparently _only_ happens in the web version?

The crash happened because the web version limits the number of validation messages to 16384 and the GLB parser didn't have the proper exception handler for that. The crash should be reproducible with the desktop tool when given a config file. Adding integration tests is planned but not as a part of this release.

emackey:
I haven't had a chance to test this one with my usual tests, but it looks like there are no major changes here, so I'm not expecting any problems from this.  Thanks!

PR #190, opened 2022-06-16
Fix AnimationDuplicateTargets

06wj:
When the animationTarget's node is undefined, it shouldn't check for the same.


06wj:
[testAnim.gltf.zip](https://github.com/KhronosGroup/glTF-Validator/files/8916724/testAnim.gltf.zip)
Add test gltf file.

image

The spec: https://www.khronos.org/registry/glTF/specs/2.0/glTF-2.0.html#reference-animation-channel-target
image

PR #191, opened 2022-06-17
Version 2.0.0-dev.3.9

lexaknyazev:
### Bugfixes

* Fixed false positive `ANIMATION_DUPLICATE_TARGETS` caused by animation targets without `node` (#190).

* Fixed false positive `ANIMATION_SAMPLER_OUTPUT_ACCESSOR_INVALID_COUNT` caused by animation targets with unknown `path`.

---

/cc @hybridherbst

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## 2.0.0-dev.3.9
+
+### Bugfixes
+
+* Fixed false positive `ANIMATION_DUPLICATE_TARGETS` caused by animation targets without `node` (#190).
+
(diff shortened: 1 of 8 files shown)


emackey (commented):
This all looks good to me.  Question, when the `path` is a known value but the `node` is undefined, should that be a warning (or error?)

```
    "animations": [
        {
            "channels": [
                {
                    "sampler": 0,
                    "target": {
                        // no "node" defined
                        "path": "rotation"
                    }
                },
```

emackey (approved):

lexaknyazev:
> This all looks good to me. Question, when the `path` is a known value but the `node` is undefined, should that be a warning (or error?)

When there are also no extensions, I think this should be a warning.

06wj:
This way of merging wipes out my code contributions.

PR #205, opened 2023-04-28
feat: add only issues option (opposite of ignoredIssues)

Apidcloud:
This PR makes it possible to provide a list of the only errors we want to validate (opposite of ignoredIssues). In my case I only want to check for 'ACCESSOR_INVALID_FLOAT', so this is easier than providing every other error using ignoredIssues.

Why? Our users often download broken models from the internet, and yet most 3d model viewers are sensible to some of the errors and can still load them. We also run them through Draco compression, and some of the errors end up not having a visible effect. But because of this issue with NaN in Draco (https://github.com/google/draco/issues/998), we want to avoid the bus error (and thus the hanging). Ideally, it is something that will be handled by Draco itself in the future; But until then, the glTF-Validator is the best tool to check and avoid such specific error.


lib/src/context.dart
```diff
@@ -25,14 +25,18 @@ import 'package:gltf/src/gl.dart' as gl;
 class ValidationOptions {
   final int maxIssues;
   final Set<String> ignoredIssues = <String>{};
+  final Set<String> onlyIssues = <String>{};
   final Map<String, Severity> severityOverrides;
 
   ValidationOptions(
-      {int maxIssues, List<String> ignoredIssues, this.severityOverrides})
+      {int maxIssues, List<String> ignoredIssues, List<String> onlyIssues, this.severityOverrides})
```
lexaknyazev, line 32: Please run `dart format .` to wrap long lines.

node/README.md
```diff
@@ -76,4 +77,4 @@ validator.validateBytes(new Uint8Array(asset), {
 });
 ```
 
-## API
+## API
```
lexaknyazev, line 80: Please keep the new line at end of file.

lib/src/context.dart
```diff
@@ -223,6 +227,10 @@ class Context {
       return;
     }
 
+    if (options.onlyIssues.length > 0 && !options.onlyIssues.contains(issueType.code)){
```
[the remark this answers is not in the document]
lexaknyazev: Sorry, this should have been `.isNotEmpty` (without brackets).

Apidcloud:
Any estimate on when this might be released?

t3h32eKK:
I have the same issue when will this be released?

t3h32eKK:
@lexaknyazev is there any chance that this will be merged soon?

PR #216, opened 2024-05-11
Update extension prefix list

emackey:


lexaknyazev:
Instead of further expanding this list, let's drop it together with the related warning for the following reasons:
- The original idea to use the validator to get more people into the extension development ecosystem didn't go as planned. Most of the registered vendor prefixes have no associated public extensions and some of them refer to entities that no longer exist.
- Since there's no fixed release schedule, this list may be out-of-sync for quite a long time thus increasing the confusion. It's not justified to force package releases on the list updates.
- Vendor extensions won't be supported anyway so the prefix-related warning makes no real difference wrt asset's validity, which is the primary purpose of the tool.

PR #230, opened 2024-08-07
Initial KHR_animation_pointer support

lexaknyazev:


emackey (approved):
I built this branch yesterday and tested it successfully with VSCode integration, no problems. As noted in source code comments here, this does leave work for the future: Validating pointers, target uniqueness, and output compatibility.  Even so, this is an improvement and can be merged.

PR #236, opened 2024-09-05
Web: Support resources in dropped directories

### LINKED ISSUE #232: Web version does not support external resources in subfolders
This applies both to the "Select" link and to the drag-n-dropped files.

Split from https://github.com/KhronosGroup/glTF-Validator/issues/229#issuecomment-2289744430.

lexaknyazev:
Fixes #232.

Notes:
- Directory access is not available when the Validator is opened from a local disk, i.e., with the `file://` URI. This is a Web platform limitation.
- Directories cannot be selected from the system dialog opened by the "Select" link. This is also a web platform limitation.
- Dropping a single directory with a glTF (or GLB) asset inside is intentionally not supported.

/cc @emackey @javagl @echadwick-artist

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -50,8 +50,10 @@
 </div>
 <div id="outputZone">
     <p><em>Validation is performed locally in your browser. Submitted assets are not uploaded.</em></p>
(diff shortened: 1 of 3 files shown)


lexaknyazev:
This is correct.

PR #247, opened 2025-12-26
Change skinned mesh warning to only consider transformed parent nodes

aaronfranke:
This PR replaces `NODE_SKINNED_MESH_NON_ROOT` with `NODE_SKINNED_MESH_PARENT_TRANSFORMS` to avoid false positives for valid models. Fixes https://github.com/KhronosGroup/glTF-Validator/issues/133

I believe this check more accurately reflects the purpose of the original check. It's not that parent-child relationships entirely are considered invalid, rather that the transform aspect of skinned mesh nodes is ignored with respect to the visible mesh. The test model `ignored_parent_transform.gltf` is already set up this way, testing with a parent translated by `[1, 2, 3]`. There are valid use cases for inserting a skinned mesh into the hierarchy, for purposes like visibility, but we only need to validate for transforms.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+### New Features
+
+* Added new `NODE_SKINNED_MESH_PARENT_TRANSFORMS` validation warning.
+
+### Changes
+
+* Changed `NODE_SKINNED_MESH_NON_ROOT` default severity to Information.
+
 ## 2.0.0-dev.3.10
 
 ### New Features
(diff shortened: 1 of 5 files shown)


aaronfranke:
Done, the original message now has reduced severity, and a different message replaces it when the parent is transformed.

lexaknyazev:
Please don't replace the original message when the new message is added. Seeing both messages upfront would be a better UX than seeing the parent transforms warning, fixing it, and discovering that the asset still has issues.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+* Added `NODE_SKINNED_MESH_PARENT_TRANSFORMS` warning which replaces some instances of `NODE_SKINNED_MESH_NON_ROOT`.
+* Reduce `NODE_SKINNED_MESH_NON_ROOT` severity to avoid unnecessary warnings for valid models.
```
lexaknyazev: Please replace with:
```
### New Features

* Added new `NODE_SKINNED_MESH_PARENT_TRANSFORMS` validation warning.

### Changes

* Changed `NODE_SKINNED_MESH_NON_ROOT` default severity to Information.
```

lexaknyazev (approved):

PR #248, opened 2025-12-28
Add full support for KHR_animation_pointer validation

### LINKED ISSUE #198: Support for KHR_animation_pointer
While Gestaltor, Babylon and Needle Engine (based on three.js) have already implemented _KHR_animation_pointer_, the validator currently doesn't support it. That leads to thousands and sometimes hundreds of thousands of errors for files that work fine:
image

The errors are many times
```json
           {
                "code": "VALUE_NOT_IN_LIST",
                "message": "Invalid value 'pointer'. Valid values are ('translation', 'rotation', 'scale', 'weights').",
                "severity": 1,
                "pointer": "/animations/0/channels/2/target/path"
            },
```

and
```json
            {
                "code": "ANIMATION_DUPLICATE_TARGETS",
                "message": "Animation channel has the same target as channel 353.",
                "severity": 0,
                "pointer": "/animations/2/channels/9/target"
            },
```

Are there plans / what are the plans for supporting _KHR_animation_pointer_ to ensure files are correctly set up? I understand that this is kind of a chicken-and-egg situation between the extension not being ratified (yet) and the need to produce correct files for it, but wanted to ask nonetheless :)

Thanks!

aaronfranke:
Implements and closes #198.

This PR finishes the work that PR #230 started, providing full support for validating `KHR_animation_pointer`, including:

- Checking that the parent properties exist in the JSON.
- Checking that the component type and vector size match the expected ones.
- Support for all mutable glTF core pointers (the read-only ones are not included).
- Allowing extensions to integrate extension pointer validation.
- Integration for all extensions already in the validator, including the KHR_texture_transform interactions, which is everything in ObjectModel.adoc except for the `EXT_lights_*` extensions and the `ADOBE_materials_*` extensions.
- Test files for all of the above, including both positive and negative tests.


lexaknyazev:
@aaronfranke I respect the efforts but certainly do not agree with the proposed implementation design. Going forward, please discuss non-trivial changes before proposing them.

aaronfranke:
@lexaknyazev What is the alternative implementation design you propose?

lexaknyazev:
The implementation proposed in this PR has multiple issues such as: hard-coded and duplicated logic, suboptimal use of the validator's API, several unreachable and untested sections, etc.

I have a quite different design drafted but not yet finished. May I borrow a few bits from this PR? This would speed up finishing that work.

aaronfranke:
@lexaknyazev You may grab any code you want from this PR, or any future PR I make. Consider my work public domain.

